PR #87, opened 2021-07-18
Do not export PROMPT_COMMAND on bash

jvmakine:
No point of exporting `PROMPT_COMMAND` as the function `change_hermit_env` won't be in the subshells anyway.


alecthomas (approved):

camh-:
There's a few of them in `shell/files/activate.tmpl/sh` too: `git grep "export PROMPT_COMMAND"` will show them.

PR #94, opened 2021-07-22
Run a manifest validation as a part of "hermit test" to detect problems in all architectures

jvmakine:
This only checks that `source` is defined for all "leaf" blocks. We can add further validations later.


jvmakine:
Disregard for now, I still need to handle the autoversioning

PR #111, opened 2021-09-06
Use core platforms when validating manifests

jvmakine:
This change means that manifests only supporting `linux-arm64` will not pass the validation. It unifies the core packages across validations, though.


alecthomas (approved):
I think this is preferable. Most packages currently only support linux/amd64, darwin/amd64 and darwin/arm64, and I suspect quite a few don't even support darwin/arm64 despite having a `source` for it. Support for linux/arm64 is mostly just sheer luck I suspect.

Related, I was thinking that maybe as part of the validation we should HEAD the URLs to be sure they exist?

jvmakine:
I love the idea for HEADing the URLs on validation. Here is a ticket for that https://github.com/cashapp/hermit/issues/112

PR #283, opened 2022-06-01
fix: Do not remove source dir from disk if both pull and clone fail

jvmakine:
Turns out that we were removing the source directory before running `git clone` If that failed, the directory was removed, leading to potentially wrong manifests being used from separate sources.

Fixes https://github.com/cashapp/hermit/issues/282


alecthomas (approved):
Nice catch!

PR #326, opened 2022-10-11
Note generated scripts

quad:
Add comments to `hermit` and `activate-hermit` to note they're generated scripts that shouldn't be modified.

A teammate made a PR to change the `hermit` script. I suspect a note like this would be helpful for Future Generations.


alecthomas (approved):
I think this will need a similar kind of change done for the internal version.

quad:
> I think this will need a similar kind of change done for the internal version.

I don't know how to do that. 🤷

syncom:
> > I think this will need a similar kind of change done for the internal version.
> 
> I don't know how to do that. shrug

IIRC, this should happen automatically for the other version :zipper_mouth_face:

PR #377, opened 2023-05-31
ops: don't re-apply already applied operations

quad:
_(This solution is _horrible_ as it interacts with the `HERMIT_ENV_OPS` envar setup in the activate script. I'd appreciate suggestions for a better approach!)_

Hermit environment variable operations can be overrode by the user after
an environment has already been opened. If this is the case, then we
shouldn't re-apply the operations; we should instead expressed respect
the wish of a mangled environment by the user.

This is relevant for constructs like Python virtual environments.

Context: https://square.slack.com/archives/C01RTEH4MFV/p1685522543591719

### DIFF
--- a/env_test.go
+++ b/env_test.go
@@ -5,7 +5,6 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
-	"reflect"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


quad:
Closed in favour of #250.

PR #397, opened 2024-03-13
fix: manual environment switching with overlapping environments

jvmakine:
Previously, we deactivated previous environment after activating a new one. This meant that if both evironments shared env variables, the deactivation script would unset them for the new environment, leaving them unset. This was not issue when using shell-hooks, as those explicitly deactivate an environment before activating a new one.

For example:
```
> cat env1/bin/hermit.hcl
env = {
  FOO:BAR,
}
> cat env2/bin/hermit.hcl
env = {
  FOO:BAR,
}
> cd env1
> . bin/activate-hermit
Hermit environment /Users/juho/tmp/hermit-test/env1 activated
> echo $FOO
BAR
> cd ../env2
> . bin/activate-hermit
Hermit environment /Users/juho/tmp/hermit-test/env1 deactivated
Hermit environment /Users/juho/tmp/hermit-test/env2 activated
> echo $FOO

>
```


shell/posix.go
```diff
@@ -3,8 +3,8 @@ package shell
 import (
 	_ "embed" // Embedding files.
 	"fmt"
-	"html/template"
 	"io"
+	"text/template"
```
jvmakine, line 7: `html/template` was escaping single quotes, so switched to `text/template`

alecthomas (approved):
Nice catch.

PR #413, opened 2024-06-21
docs: Add github-token-auth documentation

abhinav:
PR #409 added support for opting packages into GitHub token
authentication with a `github-token-auth` block in `bin/hermit.hcl`.

This adds information about this feature to the private releases
documentation, and adds an entry to the config reference.

---

This should probably not be merged until the next release is tagged
to prevent incorrect documentation from being published.


abhinav:
Superseded by #449

PR #423, opened 2024-12-07
fix(flock): race condition in acquire

nickajacks1:
There was a race condition that enabled multiple processes to acquire the same lock file. Consider the following sequence:

- Process A successfully acquires lock
- Process B enters acquire function
- Process B calls unix.Open
- Process A calls release(), deleting the lock file
- Process B calls unix.Flock on the still-open file descriptor, successfully locking the now deleted file.
- Process C successfully acquires lock on newly created lock file

After this sequence both B and C hold the lock. To resolve this race, we can simply avoid removing the file. This is what e.g. dpkg does: the /var/lib/dpkg/lock file is always present under normal circumstances.

The issue was discovered by repeatedly running TestContinueFlock:
```
=== RUN   TestContinueFlock
    flock_test.go:85: Did not expect an error but got:
        remove /tmp/TestContinueFlock3236442975/001/lock: no such file or directory
--- FAIL: TestContinueFlock (0.10s)
```
That error, while otherwise benign, indicates that the two goroutines in the test hit the race condition in question.

### DIFF
--- a/util/flock/flock.go
+++ b/util/flock/flock.go
@@ -32,6 +32,9 @@ var getPID = os.Getpid
 // updated.
 //
 // If the lock is held by another process, Acquire will block until the lock is released or the context is cancelled.
+//
+// The file is NOT deleted on release; doing so creates a race condition that allows multiple processes to acquire
+// the same lock.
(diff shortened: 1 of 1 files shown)


alecthomas (approved):
Nice catch!

PR #427, opened 2025-01-03
fix: env validation when missing fish script

mightyguava:
env validation was failing when the `activate-hermit.fish` script was missing due to matching on path instead of filename. Not sure why integration tests didn't catch it. Will follow up later with a test.


staktrace (approved):

mightyguava:
@staktrace do you know who is allowed to approve…?

jvmakine (approved):

PR #438, opened 2025-02-23
feat: GitHub CLI authentication support

lox:
Adds support for using GitHub CLI (`gh`) as an authentication provider for Hermit,  eliminating the need to manually set `HERMIT_GITHUB_TOKEN`.

Exposing `HERMIT_GITHUB_TOKEN` securely across many developer environments is tricky to do safely without exposing it to other tools or distributing shell wrappers for hermit. Using `gh auth token` allows for short-lived tokens that are only exposed to hermit as needed. 

This feature is enabled via setting the following in the user hermit config (~/.hermit.hcl):

```
gh-cli-auth = true;
```

### DIFF
--- a/app/main.go
+++ b/app/main.go
@@ -20,6 +20,7 @@ import (
 	"github.com/cashapp/hermit"
 	"github.com/cashapp/hermit/cache"
 	"github.com/cashapp/hermit/github"
+	"github.com/cashapp/hermit/github/auth"
 	"github.com/cashapp/hermit/sources"
 	"github.com/cashapp/hermit/state"
 	"github.com/cashapp/hermit/ui"
@@ -170,14 +171,6 @@ func Main(config Config) {
 		cli = &unactivated{cliBase: common}
 	}
 
-	githubToken := os.Getenv("HERMIT_GITHUB_TOKEN")
-	if githubToken == "" {
-		githubToken = os.Getenv("GITHUB_TOKEN")
-		p.Tracef("GitHub token set from GITHUB_TOKEN")
-	} else {
-		p.Tracef("GitHub token set from HERMIT_GITHUB_TOKEN")
-	}
-
 	kongOptions := []kong.Option{
 		kong.Groups{
 			"env":    "Environment:\nCommands for creating and managing environments.",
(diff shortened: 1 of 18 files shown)


lox:
This was tricky to test! 

What I've done is add a build tag of `localgithub` that is used by the integration tests to compile hermit. If compiled this way, it respects `HERMIT_GITHUB_BASE_URL` as an alternative to `github.com` which allows tests to create a mock github endpoint without the horrors of HTTPS proxying.

Lemme know if this is overkill @alecthomas.

github/auth/provider.go
```diff
@@ -0,0 +1,88 @@
...
+	p.ui.Tracef("gh CLI found in PATH: %s", ghPath)
+
+	// Run gh auth token
+	cmd := exec.Command("gh", "auth", "token")
+	output, err := cmd.Output()
+	if err != nil {
+		return "", errors.Wrap(err, "failed to get token from gh CLI")
+	}
+
+	p.ui.Tracef("gh auth token output: %s", string(output))
```
nickajacks1, line 73: Hmmm, this prints a secret. It's probably better to either remove this or warn in the cli usage that trace may leak sensitive information.

lox: Good catch, was there for dev.

nickajacks1 (commented):
Seems good to me. Just had one more little comment.

app/main.go
```diff
@@ -217,13 +227,36 @@ func Main(config Config) {
...
+	var githubToken string
+	if envInfo != nil && len(envInfo.Config.GitHubTokenAuth.Match) > 0 {
+		// Determine provider based on user config
+		providerType := "env"
+		if userConfig.GHCliAuth {
+			providerType = "gh-cli"
+		}
+
+		provider, err := auth.NewProvider(providerType, p)
+		if err != nil {
```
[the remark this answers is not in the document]
lox: Fixed, both are now Fatal.

nickajacks1 (approved):
Not a maintainer but looks LGTM to me

lox:
Appreciate the review @nickajacks1

alecthomas (approved):
I like it!

PR #442, opened 2025-03-06
fix: Revert "perf(exec): improve performance"

wsutina:
## 📝 Description
In a nested environment, Hermit fails to set the correct environment for the target binary. For example, given the following directory structure:
```
repo-root/
├── bin/
│   ├── java@17
│   ├── gradle
├── foo/
│   ├── bin/
│   │   ├── java@21
```
When navigating to `repo-root/foo` and running `gradle`, Hermit installs Java 17 but sets `JAVA_HOME` to Java 21, resulting in the following error:
```
JAVA_HOME is set to an invalid directory: /root/.cache/hermit/pkg/openjdk@21
```
This issue leads to unexpected failures in internal builds.

## Revert Reason
This commit reverts cashapp/hermit#433 because it introduces a regression in environment resolution. Until a proper fix is found, reverting ensures stability for existing builds.

PR #448, opened 2025-03-08
perf(exec): improve performance

nickajacks1:
The previous attempt in 9a9737ec668f5aa01cb8a2b83d5832654d6e4997 (#433) had an invalid method of detecting when environment variables needed to be parsed, so it was reverted in cc90510721cc58923414ce1efd339b3e2ffb20b8 (#442). This attempt takes the approach of instead checking if the binary being executed is inside the currently activated environment. If not, then we load the environment variables of all packages in the exec'd binary's environment. It may be possible to get away with only parsing the package's dependencies, but I'm not confident that would correctly handle all present use cases.

### DIFF
--- a/env.go
+++ b/env.go
@@ -49,7 +49,7 @@ var (
 	//go:embed "files/install.sh.tmpl"
 	InstallerTemplateSource string
 
-	// UserStateDir should be passed to Open()/Init() in most cases.
+	// UserStateDir should be passed to OpenEnv()/Init() in most cases.
 	UserStateDir = func() string {
 		// Check if state dir is explicitly set
 		explicit := os.Getenv("HERMIT_STATE_DIR")
@@ -115,6 +115,7 @@ type Env struct {
 	httpClient      *http.Client
 	scriptSums      []string
 	sourceRewriters []sources.URLRewriter
+	activated       bool
 
 	// Lazily initialized fields
 	lazyResolver  *manifest.Resolver
@@ -353,6 +354,7 @@ func OpenEnv(
 		httpClient:      httpClient,
 		scriptSums:      scriptSums,
 		sourceRewriters: sourceRewriters,
+		activated:       os.Getenv("ACTIVE_HERMIT") == info.Root,
 	}, nil
 }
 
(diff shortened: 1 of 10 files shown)


nickajacks1:
To the best of my ability, I have determined that this attempt is _more_ correct than #433. Without more information about the internal build failures the prior PR caused, however, I cannot be fully confident. Out of caution, I have marked this as a draft.

env.go
```diff
@@ -349,6 +350,7 @@ func OpenEnv(
 		ephemeralEnvars: envars.Infer(ephemeral.System()),
 		httpClient:      httpClient,
 		scriptSums:      scriptSums,
+		activated:       os.Getenv("ACTIVE_HERMIT") == info.Root,
```
mpeyper, line 353: If I'm following correctly, the main change from before to now is that we consider the environment already activated if the currently active environment (`os.Getenv("ACTIVE_HERMIT")) matches the closest hermit root (`info.Root`), which in the case that broke before would be the nested environment, while [before](https://github.com/cashapp/hermit/pull/433/files#diff-901643e46196ff948ed22a0164ff057a898f5ec2c278edb5c8d905f369c929f0R353) we only considered it already activated if there wasn't a deactivated environment, right? This feels more correct to me.

integration/integration_test.go
```diff
@@ -279,6 +279,14 @@ EOF
...
 				assert test "$(./testenv2/bin/testbin1)" = "testbin1 1.0.1"
 			`},
+		{name: "ExecuteFromOtherEnvironmentLoadsDependentEnvars",
+			preparations: prep{allFixtures("testenv1", "testenv2")},
+			script: `
+				testenv2/bin/hermit install testbin1-1.0.0 testbin4
+				. testenv1/bin/activate-hermit
+				hermit install testbin1-1.0.1
+				assert test "$(./testenv2/bin/testbin4)" = "env[1.0.0] exec[testbin1 1.0.0]"
+			`},
```
mpeyper, line 289: I would like to see a test setup similar to this, but in a nested environment (like the `environment_inheritance` test data) as this is the scenario we saw failing (nested environment calling parent binary with a call to an overwritten dependency).

env.go
nickajacks1: Right, this is my intention. That said, I still want to scan through the code more to be certain that this is indeed the correct way to accomplish that rather than a potentially fragile indirect way (for example, I saw that OpenEnv was called where info.Root was a pkg folder in the cache, which I wasn't expecting).

nickajacks1:
Updated with the requsted nested env integration test.

wsutina:
Hi @nickajacks1, thank you for adding the test — we really appreciate it! 🙏
We’ll test this against our internal repositories to ensure there are no regressions.

It looks like there’s a merge conflict at the moment. When you have a moment, could you please update the branch? Thanks!

nickajacks1:
Thank you, I greatly appreciate your offer to regression test this.
I have fixed the merge conflict.

nickajacks1:
I realize I accidentally allowed my editor to auto-format integration_test.go.  I will undo this shortly.

wsutina (approved):
Thanks again for putting up the change!

PR #454, opened 2025-03-24
chore: Remove docs exclusion from CI runs

jrobotham-square:
The docs exclusion saves a small amount of time for a doc-only change, but at the moment, the PR never runs the checks, so doesn't get marked as ready to merge!

We could update the checks, but feels simpler to just remove the doc exclusion.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,12 +1,8 @@
 on:
   push:
-    paths-ignore:
-      - 'docs/**'
     branches:
       - master
   pull_request:
-    paths-ignore:
-      - 'docs/**'
 name: CI
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,12 +1,8 @@
 on:
   push:
-    paths-ignore:
```
wsutina, line 3: Can keep this. The Doc workflow will run it on master https://github.com/cashapp/hermit/blob/master/.github/workflows/docs.yml

jrobotham-square: done - 4258a86caf94fcaffb0b4d0e1b2035de120e0945

wsutina (approved):

PR #474, opened 2025-04-16
fix: skipping activation when already activated should exit cleanly

mpeyper:
We've had reports of some AI agents struggling when trying to activate an already activated environment because of the non-zero exit code. There reason for exiting with the 34 code seems to have been lost to the sands of time, and we now believe that there is not a significant reason not to exit cleanly in this case.

### DIFF
--- a/it/full/spec/it_spec.sh
+++ b/it/full/spec/it_spec.sh
@@ -27,10 +27,10 @@ Describe "Hermit"
         The stderr should not be blank
         The variable HERMIT_ENV should equal "$(pwd)"
       End
-      It "fails to activate twice"
+      It "cannot be activated twice"
         . ./bin/activate-hermit
         When call source ./bin/activate-hermit
-        The status should be failure
+        The status should be success
         The stderr should eq "This Hermit environment has already been activated. Skipping"
       End
       It "loads environment variables from the local hermit.hcl"
--- a/it/full/spec/it_spec.sh
+++ b/it/full/spec/it_spec.sh
@@ -27,7 +27,7 @@ Describe "Hermit"
         The stderr should not be blank
         The variable HERMIT_ENV should equal "$(pwd)"
       End
-      It "cannot be activated twice"
+      It "considers a second activation a noop"
         . ./bin/activate-hermit
         When call source ./bin/activate-hermit
         The status should be success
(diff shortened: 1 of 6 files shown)


it/full/spec/it_spec.sh
```diff
@@ -30,7 +30,7 @@ Describe "Hermit"
       It "fails to activate twice"
```
nickajacks1, line 30: This name is arguably out of date now. Maybe something along the lines of `It "does not fail when activating twice"`

alecthomas (approved):

it/full/spec/it_spec.sh
mpeyper: Good shout! Fixed.

it/full/spec/it_spec.sh
```diff
@@ -27,10 +27,10 @@ Describe "Hermit"
         The stderr should not be blank
         The variable HERMIT_ENV should equal "$(pwd)"
       End
-      It "fails to activate twice"
+      It "cannot be activated twice"
```
jrobotham-square, line 30: nit:
```suggestion
      It "considers a second activation a noop"
```

jrobotham-square (approved):

shell/files/activate.tmpl.sh
```diff
@@ -5,7 +5,7 @@ export HERMIT_ENV={{.Root}}
 if [ -n "${ACTIVE_HERMIT+_}" ]; then
   if [ "$ACTIVE_HERMIT" = "$HERMIT_ENV" ]; then
     echo "This Hermit environment has already been activated. Skipping" >&2
-    return 34
+    return 0
```
staktrace, line 8: While I'm in favor of this change, it would be good to somehow "reserve" the exit code 34 so that it's not used for any other error in the future. We have existing code in a number of places explicitly checking for this exit code to ignore it, and it would be bad if this exit code started being used for some other error in the future

mpeyper: I've added some [basic linting](https://github.com/cashapp/hermit/pull/474/files#diff-f6b20c425ee57cd5dcc96e5b486a83154b7de95233b2e0bd9b0ff26e4be6f6e1) for it across the shell scripts. It's not perfect, but should hopefully help a bit.

it/full/spec/it_spec.sh
mpeyper: Done and updated the other test name to match.

wsutina (approved):

PR #496, opened 2025-07-14
fix: Ensure ManageGit defaults to true

jrobotham-square:
There appears to be a regression introduced by https://github.com/cashapp/hermit/pull/436

`ManageGit` should be defaulting to true, but the default defined `hermit.Config` only applies when parsing from HCL.

So if you have either
- No `~/.hermit.hcl` file.
- An `~/.hermit.hcl` file without a value set for `defaults.manage-git` 

The repo will be `init`ed with `manage-git = false`

This change sets the default to true (as a starting point) and switches the command line parsing to use a pointer to allow us to distinguish between "true", "false" and "not set".

### DIFF
--- a/app/init_cmd.go
+++ b/app/init_cmd.go
@@ -7,7 +7,7 @@ import (
 )
 
 type initCmd struct {
-	Git     bool     `negatable:"" help:"Enable Hermit's automatic management of Git'"`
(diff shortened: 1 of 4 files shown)


wsutina (approved):
LGTM!

alecthomas (approved):

PR #503, opened 2025-08-20
fix: Renames if source and destination are on different filesystems

js-murph:
# What?

Has a fallback strategy for moving files when it's unable to perform an `os.Rename()`

# Why?

If the source and destination are on different mount locations then the Rename will fail with an error like...

`rename /root/.cache/hermit/pkg/hermit@square /root/.cache/hermit/pkg/.hermit@square.old: invalid cross-device link`

This has a fallback to copy the file if it fails to perform the rename.

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -294,7 +294,7 @@ func (s *State) renameRecursive(b *ui.Task, src, dest string) error {
 			if rerr != nil {
 				return errors.WithStack(rerr)
 			}
-			if werr := os.WriteFile(dest, input, 0o755); werr != nil {
+			if werr := os.WriteFile(dest, input, 0o600); werr != nil {
 				return errors.WithStack(werr)
 			}
 			return errors.WithStack(os.Remove(src))
--- a/state/state.go
+++ b/state/state.go
@@ -8,6 +8,7 @@ import (
 	"regexp"
 	"runtime"
 	"strings"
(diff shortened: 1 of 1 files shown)


state/state.go
```diff
@@ -284,7 +285,24 @@ func (s *State) renameRecursive(b *ui.Task, src, dest string) error {
 	})
 
 	task.Debugf("mv %s %s", src, dest)
-	return errors.WithStack(os.Rename(src, dest))
+
+	if err := os.Rename(src, dest); err != nil {
+		// If the rename fails with EXDEV, it means the source and destination are on different filesystems.
+		// In this case, we read the source file and write it to the destination path.
```
boblail, line 291: Do we need to remove the `src` file? (Unsure, just asking)

js-murph:
> Ah! Somewhat surprisingly, your comment in here makes more sense than the other one! 😂
> 
> So the problem is caused by Hermit trying to update itself from within Docker? Maybe we should add a flag to disable updates... or step back and think.

I find if I write a sentence enough times it will eventually make sense to someone 😂. Disabling the updates would fix this I think, I think there are also times where we've relied on that auto-updating feature to roll out fixes. I'm ok either way but it may have implications for releasing hermit changes.

Perhaps @wsutina or @mpeyper should weigh in as this affects them more.

PR #513, opened 2025-09-29
perf: speed up shell completion for install and info commands

nickajacks1:
Instead of loading every possible package, we can just load the ones that match what the user is completing.

This cuts tab completion time for `hermit install` from 10+ seconds to 0-1 seconds.


nickajacks1:
Oops, didnt mean to close this. Opened it on the wrong branch and force pushed master. New PR: #518

PR #520, opened 2025-11-21
Do not hard-code /bin/bash

JamesGuthrie:
While /bin/bash does exist on many platforms, it's not guaranteed to be the case. One prominent example is NixOS. The portable way to invoke scripts is to use #!/usr/bin/env bash which finds the most appropriate bash binary from the user's configured PATH.


JamesGuthrie:
> I'm very curious why you'd use Hermit with NixOS...Nix seems like a superset of Hermit.

I use NixOS, but the wider team I work with doesn't - they use Hermit. It's easier for me to adjust to them than for them to adjust to me - aside from the hardcoded /bin/bash paths.

JamesGuthrie:
@alecthomas are you in principle open to this change?

PR #550, opened 2026-03-12
perf: avoid dirname in activation shell hooks

nickajacks1:
Repeated forking of a dirname process can get rather heavy in deep directory structures, especially with distributions that are moving to uutils. This is most problematic for bash, which runs PROMPT_COMMAND for every prompt as opposed to zsh and fish which run the activation check on chdir. Switching to a pure shell calculation of the parent directory makes performance scale much better.

On an Intel Core Ultra 7 165H with uutils, the existing dirname solution takes about 30 milliseconds to traverse 10 directories, which is enough to be noticeable. Using the updated pure shell approach reduces the time down to 5-10 milliseconds at a directory depth of 80.


alecthomas (approved):

alecthomas:
I think this has been mitigated somewhat by changes to when hooks run - some of them only run on working directory change now.

nickajacks1:
@alecthomas  Sadly, the recursive dirname search still executes on every prompt. This is because it uses PROMPT_COMMAND, and bash does not really have any "on chdir" type hook. (To be honest, it's still icky and noticeable even on cd). Would you consider reopening this?

PR #558, opened 2026-04-15
feat: parrallellize hermit install downloads

stuartwdouglas:
Download package archives concurrently during install using errgroup, with concurrency defaulting to `runtime.NumCPU()`. A new `-j` flag allows overriding the parallelism level.

Only the download phase is parallelised. Extraction and linking remain serial because unpack triggers may execute binaries from dependency packages.

Both install paths benefit:
- **No-args** (re-install existing): parallel download, then serial CacheAndUnpack.
- **With packages**: parallel download, then serial install/link loop.

A new `State.Download` method is added to expose download-only functionality, separate from `CacheAndUnpack`.

### DIFF
--- a/app/log_test.go
+++ b/app/log_test.go
@@ -33,11 +33,11 @@ func TestLogs(t *testing.T) {
 			assert.NoError(t, err)
 		},
 		tmpl: `
+			debug:tpkg-0.9.0:download: Downloading {{.Source}}
+			debug:tpkg-0.9.0:unpack: Extracting {{.Cache}} to {{.State}}/pkg/tpkg-0.9.0
 			info:tpkg-0.9.0:install: Installing tpkg-0.9.0
 			debug:tpkg-0.9.0:install: From {{.Source}}
 			debug:tpkg-0.9.0:install: To {{.State}}/pkg/tpkg-0.9.0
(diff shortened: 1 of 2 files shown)


alecthomas (approved):

app/log_test.go
```diff
@@ -33,11 +33,11 @@ func TestLogs(t *testing.T) {
 			assert.NoError(t, err)
 		},
 		tmpl: `
+			debug:tpkg-0.9.0:download: Downloading {{.Source}}
+			debug:tpkg-0.9.0:unpack: Extracting {{.Cache}} to {{.State}}/pkg/tpkg-0.9.0
 			info:tpkg-0.9.0:install: Installing tpkg-0.9.0
 			debug:tpkg-0.9.0:install: From {{.Source}}
 			debug:tpkg-0.9.0:install: To {{.State}}/pkg/tpkg-0.9.0
-			debug:tpkg-0.9.0:download: Downloading {{.Source}}
```
alecthomas, line 39: Is this right?

I think the reason it changed is because the order is now non-deterministic. You might need to set parallelism to 1 to make the tests deterministic.

stuartwdouglas: Its correct, it parallelized the download, but then does the install in a loop. There is only one package in this test.

PR #559, opened 2026-04-16
Add install-on-activate to eagerly install selected packages on env activation

joshfriend:
Hermit packages are normally downloaded on-demand when a stub binary in `./bin` is first invoked. This causes problems for packages like openjdk that export environment variables pointing to the package root (e.g. `JAVA_HOME`). Tools like Gradle, Android Studio, and `gradlew` read `JAVA_HOME` directly and run `$JAVA_HOME/bin/java` without going through hermit stubs — if the JDK hasn't been downloaded yet, that path doesn't exist and the tools fail.

This adds an `install-on-activate` list to `bin/hermit.hcl` that specifies which packages should be eagerly downloaded and unpacked when the environment is activated:

```hcl
install-on-activate = ["openjdk"]
```

Only the listed packages are installed eagerly; everything else remains lazy. This keeps activation fast for environments with many packages while ensuring critical ones like the JDK are always present.


alecthomas (approved):
Glorious!

PR #567, opened 2026-05-19
fix: reject invalid env-var names from manifests

alecthomas:
Manifest env = { ... } keys were interpolated unquoted into the
activation scripts and ApplyEnvars output, allowing a malicious
package to inject shell commands that run on activation and re-fire
on every prompt redraw via update_hermit_env.

Validate keys against the POSIX identifier syntax
^[A-Za-z_][A-Za-z0-9_]*$ at every trust boundary (manifest resolve,
hermit.hcl load, --env CLI flag) and again at emit time in ApplyEnvars
and ActivateHermit as defence in depth.

Co-authored-by: Claude Code <noreply@anthropic.com>

### DIFF
--- a/env.go
+++ b/env.go
@@ -287,6 +287,9 @@ func readConfig(configFile string) (*Config, error) {
 		if err != nil {
 			return nil, errors.Wrap(err, configFile)
 		}
+		if err := config.Envars.Validate(); err != nil {
+			return nil, errors.Wrap(err, configFile)
+		}
 	}
 	return config, nil
 }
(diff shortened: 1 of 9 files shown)


jrobotham-square (approved):
nit: This will fix the vulnerability, but if users have a dodgy value in their `hermit.hcl` they won't be able to use hermit itself to remove it.  eg `hermit env --unset 'FOO-BAR'`

manual fix is just to edit the hcl... which is pretty easy given the error message, so probably fine.

PR #573, opened 2026-06-15
fix: confine archive extraction within the destination directory

jrobotham-square:
Extraction now performs all filesystem operations through `os.Root`, so symlinks contained in an archive can no longer cause files to be written outside the destination directory. Also avoids following symlinks during the read-only finalisation chmod, and uses a fixed name for the nested payload when unpacking `.deb` packages.

Adds regression tests for symlink handling during extraction.

### DIFF
--- a/archive/archive.go
+++ b/archive/archive.go
@@ -411,7 +411,7 @@ func extractZip(b *ui.Task, f *os.File, info os.FileInfo, dest string, strip int
 			return errors.Wrap(err, destFile)
 		}
 	}
-	return nil
+	return validateSymlinks(root)
 }
 
 func extractZipFile(root *os.Root, zf *zip.File, destFile string, dest string) error {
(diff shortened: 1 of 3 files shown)


jrobotham-square:
@codex review

archive/archive.go
```diff
@@ -684,3 +703,72 @@ func sanitizeSymlinkTarget(destFile, linkTarget, destination string) error {
...
+		if d.Type()&fs.ModeSymlink == 0 {
+			return nil
+		}
+		if symlinkEscapesRoot(root, name) {
+			target, _ := root.Readlink(name)
+			return errors.Errorf("%s: illegal symlink target %q (resolves outside extraction root)", name, target)
+		}
+		return nil
+	})
+}
```
alecthomas, line 754: Do we even need to do this if we're using os.Root()? Is it even possible to symlink outside the root?

alecthomas (approved):

archive/archive.go
alecthomas: I looked this up and yes, yes we do.

jrobotham-square: yep. can confirm:
> os.Root refuses to follow a symlink out of the root during OpenFile/Stat/etc. But os.Root.Symlink does not validate the target — it happily creates a link whose target points outside the root.

PR #586, opened 2026-07-25
fix: show Fish activation script after init

Noot-tooN:
## Summary

- detect the current shell when a Hermit environment is initialized
- recommend `activate-hermit.fish` when Fish is detected
- preserve the existing activation command for Bash, Zsh, and detection failures

## Why

Hermit generates both `activate-hermit` and `activate-hermit.fish`, but the
post-`init` message always recommended the Bash/Zsh script. Following that
instruction from Fish causes Fish to parse an incompatible shell script.

The change reuses Hermit's existing shell detection and only changes the
recommended script when Fish is detected.

## Testing

- `go test ./...`
- `git diff --check`

### DIFF
--- a/env.go
+++ b/env.go
@@ -201,13 +201,14 @@ func Init(l *ui.UI, env string, distURL string, stateDir string, config Config, 
 			}
 		}
 	}
+	detectedShell, _ := shell.Detect()
 	l.Infof(`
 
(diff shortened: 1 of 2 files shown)


alecthomas (approved):

alecthomas:
Good idea!

PR #587, opened 2026-07-25
Resolve Hermit binary from PATH in shell hooks

Noot-tooN:
## Summary

- resolve the installed `hermit` executable from `PATH` when generating Bash, Zsh, and Fish shell hooks
- quote the resolved executable path so installations in paths containing spaces work
- execute the generated snippets in real Bash, Zsh, and Fish processes during tests
- install Fish and Zsh in the CI test job so every supported shell is exercised

## Why

The generated shell-hook configuration assumed Hermit was installed at `$HOME/bin/hermit`. Installations managed elsewhere, such as Homebrew at `/opt/homebrew/bin/hermit`, therefore produced an inactive hook: entering a Hermit environment did not update `PATH`, so project tools such as `npx` remained unavailable.

The hook installer now locates the same `hermit` executable already available on `PATH` while preserving an explicitly configured `HERMIT_ROOT_BIN`.

## Validation

- `go test -count=1 ./...`
- `./bin/golangci-lint run`
- `./bin/lint-shell-scripts`
- `git diff --check`


.github/workflows/ci.yml
```diff
@@ -15,6 +15,8 @@ jobs:
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
       - name: Init Hermit
         uses: cashapp/activate-hermit@cea9af7913204a965fd488637a8d1811bba2e616 # v1
+      - name: Install shells
+        run: sudo apt-get install --no-install-recommends fish zsh
```
Noot-tooN, line 19: Not sure how the maintainers of this repo would feel about this, but i don't see any proper way to test this feature out without this. 😄

PR #594, opened 2026-07-27
fix: re-assert Hermit PATH precedence at first zsh prompt

boblail:
Since #554, zsh hooks activate Hermit eagerly, so when the hooks are installed early (eg. in /etc/zshrc) anything the user's startup files later prepend to PATH (eg. RVM) shadows the Hermit environment in any terminal opened inside it. This registers a one-shot precmd that moves Hermit's PATH entries back to the front at the first prompt — the same deferral `update_hermit_ps1` already does for the prompt prefix. `HERMIT_PREPEND_PATH` components stay ahead of the environment's bin, and later prepends are demoted rather than removed. zsh only; bash hooks already run at prompt time.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -683,6 +683,41 @@ EOF
 			`,
 		},
 		{
+			// Hermit hooks can activate an environment before the user's own
+			// startup files run (eg. from /etc/zshrc), letting anything they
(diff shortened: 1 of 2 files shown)


alecthomas (approved):
I think this makes a lot of sense

PR #598, opened 2026-08-05
Allow packages without binaries or apps when they provide trigger actions or env vars

joahg:
Context: #597 — composing agent-skills-style content packages from existing Hermit primitives.

Hermit currently rejects any package manifest that does not declare at least one binary or app. This prevents composing content-only packages (e.g. a git-sourced payload that is materialised into the environment via `on unpack`/`on activate` actions) without adding a dummy no-op binary that pollutes `bin/` and `PATH`.

This relaxes the validation so a package with no binaries and no apps is still valid when it declares at least one resolved lifecycle trigger action or environment variable. Empty trigger blocks such as `on unpack {}` do not count, and manifests that provide none of these remain rejected with the existing `ErrNoBinaries` error.

Integration tests exercise content-only installation and reactivation, environment-only packages, and rejection of empty trigger-only manifests. All cases run through the existing `./integration` harness against a freshly built CLI in Bash and Zsh. The new regression cases were also checked against current master without this PR’s implementation and failed at the behavior this PR fixes.

Validation: `bin/go test ./...`, `bin/go test -count=1 -tags integration ./integration -run '^TestIntegration$'`, and `bin/golangci-lint run`. The separate Fish integration test is not included locally because Fish is unavailable.

Tracking: AGNTOPS-430.


manifest/resolver.go
```diff
@@ -519,8 +519,16 @@ func newPackage(manifest *AnnotatedManifest, config Config, selector Selector) (
 			files[k] = v
 		}
 	}
-	// Verify.
-	if len(p.Binaries) == 0 && len(p.Apps) == 0 {
+	// Verify. Packages without binaries or apps are permitted as long as they
```
alecthomas, line 523: "Verify" this by writing an integration test please, in `./integration`.

alecthomas (approved):
Can you add an integration test with one of these new manifest types? The integration

PR #601, opened 2026-08-05
Run the full activation lifecycle in 'hermit env --activate'

joahg:
Context: #597 — composing agent-skills-style content packages from existing Hermit primitives.

`hermit activate` installs `install-on-activate` packages and runs `on activate` triggers before emitting environment variables, but `hermit env --activate` — used by the shell hooks in `activate.tmpl.sh`/`activate.tmpl.fish` when the environment changes, and commonly by CI via `eval $(hermit env --activate)` — only computed and applied the envar operations. The two activation paths could therefore drift: packages and triggers were skipped entirely on the `env --activate` path.

This extracts the shared lifecycle into `Env.Activate` and uses it from both paths. Trigger messages are printed to stderr on the `env --activate` path since stdout is reserved for shell commands.

An integration test starts with installed stubs and an empty package cache, verifies that env --activate unpacks the package and runs its trigger, checks emitted shell commands and stderr messages, and verifies that env --ops has no activation side effects. All cases run through the existing `./integration` harness against a freshly built CLI in Bash and Zsh. The new regression cases were also checked against current master without this PR’s implementation and failed at the behavior this PR fixes.

Validation: `bin/go test ./...`, `bin/go test -count=1 -tags integration ./integration -run '^TestIntegration$'`, and `bin/golangci-lint run`. The September 18 refresh also passed the full Go integration suite, including Fish, plus shell lint.

Tracking: AGNTOPS-430.

September 18 refresh: merged current master without rewriting history. Resolved overlapping lifecycle and channel-freshness test insertions by retaining both. Full unit tests, full Go integration tests (Bash, Zsh, and the Fish regression), golangci-lint, and shell lint passed at https://github.com/joahg/hermit/commit/7011a77.


alecthomas (approved):

joahg:
🤖 Joah's AI agent: refreshed this branch against current master at https://github.com/joahg/hermit/commit/7011a77. The conflict was overlapping independent test additions; both are preserved. Full unit tests, full Go integration tests (Bash/Zsh plus Fish regression), golangci-lint, and shell lint pass locally. A delta review is underway.

@alecthomas could you approve the fresh fork workflow run https://github.com/cashapp/hermit/actions/runs/35371637691 ? It is action_required with no jobs started. Joah has requested landing this PR once the delta review and fresh CI are clear; this account lacks upstream merge permission.

PR #603, opened 2026-08-11
Disable curlrc invocation for installation

richardchesse:
The installer runs `curl | bash` which is generally frowned upon for several reasons. One is that it takes the responses directly from curl to bash.
This fails if the user has any HTTP diagnostic output in their curlrc, as the output is interpreted by the gzip command. This PR disables curlrc
from being sourced, working around the issue. It's not as elegant as downloading the installer, then invoking bash, but it works.


alecthomas (approved):

alecthomas:
Disabling curlrc for every consumer seems like a pretty heavy hammer to work around using `-w`.

PR #604, opened 2026-08-11
Download installer archive before decompression

richardchesse:
## Summary

- download the compressed Hermit binary to a temporary archive before decompressing it
- keep user curl configuration enabled without allowing `--write-out` diagnostics to corrupt the archive stream
- remove partial downloads on curl errors and reject missing or empty archives
- keep both embedded installer templates synchronized

## Why

The installer previously piped curl's stdout directly into gzip. If a user's curlrc configured diagnostic output with `--write-out`, those bytes were mixed into the archive stream and decompression failed. Disabling curlrc globally, as proposed in #603, avoids the symptom but also ignores every user's curl configuration. Downloading to a file isolates the response body while preserving curlrc behavior.

## Validation

- `./bin/go test ./...`
- generated-installer parity check
- Shellcheck on the generated installer
- regression harness with curlrc `--write-out` output
- rejection test for a zero-byte download

### DIFF
--- a/files/install.sh.tmpl
+++ b/files/install.sh.tmpl
@@ -61,9 +61,16 @@ function user_install() {
   fi
   URL="${HERMIT_DIST_URL}/hermit-${OS}-${ARCH}.gz"
   TMP_FILE="${HERMIT_EXE}.download.${RANDOM}"
+  TMP_ARCHIVE="${TMP_FILE}.gz"
   echo "Downloading ${URL} to ${HERMIT_EXE}"
   chmod -f u+w "${HERMIT_EXE}" 2> /dev/null || true
-  curl -fsSL "${URL}" | gzip -dc > "${TMP_FILE}"
+  curl --remove-on-error -fsSL -o "${TMP_ARCHIVE}" "${URL}"
+  if [ ! -f "${TMP_ARCHIVE}" ] || [ ! -s "${TMP_ARCHIVE}" ]; then
+    echo "Failed to download ${URL}" >&2
+    exit 1
+  fi
+  gzip -dc "${TMP_ARCHIVE}" > "${TMP_FILE}"
+  rm -f "${TMP_ARCHIVE}"
   chown "$ID_USER:$ID_GROUP" "${TMP_FILE}"
   chmod u+wx "${TMP_FILE}"
   mv "${TMP_FILE}" "${HERMIT_EXE}"
--- a/cmd/geninstaller/install.sh.tmpl
+++ b/cmd/geninstaller/install.sh.tmpl
@@ -61,9 +61,16 @@ function user_install() {
   fi
   URL="${HERMIT_DIST_URL}/hermit-${OS}-${ARCH}.gz"
   TMP_FILE="${HERMIT_EXE}.download.${RANDOM}"
+  TMP_ARCHIVE="${TMP_FILE}.gz"
   echo "Downloading ${URL} to ${HERMIT_EXE}"
   chmod -f u+w "${HERMIT_EXE}" 2> /dev/null || true
-  curl -fsSL "${URL}" | gzip -dc > "${TMP_FILE}"
+  curl --remove-on-error -fsSL -o "${TMP_ARCHIVE}" "${URL}"
+  if [ ! -f "${TMP_ARCHIVE}" ] || [ ! -s "${TMP_ARCHIVE}" ]; then
+    echo "Failed to download ${URL}" >&2
+    exit 1
+  fi
+  gzip -dc "${TMP_ARCHIVE}" > "${TMP_FILE}"
+  rm -f "${TMP_ARCHIVE}"
   chown "$ID_USER:$ID_GROUP" "${TMP_FILE}"
   chmod u+wx "${TMP_FILE}"
   mv "${TMP_FILE}" "${HERMIT_EXE}"


files/install.sh.tmpl
```diff
@@ -61,9 +61,16 @@ function user_install() {
   fi
   URL="${HERMIT_DIST_URL}/hermit-${OS}-${ARCH}.gz"
   TMP_FILE="${HERMIT_EXE}.download.${RANDOM}"
+  TMP_ARCHIVE="${TMP_FILE}.gz"
   echo "Downloading ${URL} to ${HERMIT_EXE}"
   chmod -f u+w "${HERMIT_EXE}" 2> /dev/null || true
-  curl -fsSL "${URL}" | gzip -dc > "${TMP_FILE}"
+  curl --remove-on-error -fsSL -o "${TMP_ARCHIVE}" "${URL}"
```
alecthomas, line 67: 🤖 **Blocking:** `--remove-on-error` was added in curl 7.83.0 (May 2022), so this breaks installation on every platform with an older curl. Curl exits with code 2 on an unknown option *before making any request*, and under `set -euo pipefail` that aborts the whole install.

I verified this with a stub emulating curl 7.81: the previous script installs successfully, while this one prints `curl: option --remove-on-error: is unknown` and exits 2 with nothing installed.

Affected platforms include Ubuntu 22.04 LTS (curl 7.81.0), Ubuntu 20.04 (7.68.0), Debian 11 (7.74.0), RHEL/Rocky 8 (7.61.1) and RHEL 9 (7.76.1), and macOS 12 and earlier (7.79.1 or older). The script's only stated curl requirement is that curl exists at all (line 24), and no minimum version is documented. This is both the documented `curl ... install.sh | bash` bootstrap path and the self-healing path baked into every generated `bin/hermit` stub (line 113), so affected users lose fresh installs and auto-bootstrap in existing repos, with an error message that gives no hint about the cause.

CI will not catch this, because `ubuntu-latest` and `macos-latest` both ship curl 8.x.

The flag is also not load-bearing: with `-f` plus `set -e`, a failed download already aborts the script, and `--remove-on-error` only deletes the partial file. A `trap` does the same thing portably and also fixes the leftover-temp-file problem below:

```bash
trap 'rm -f "${TMP_ARCHIVE}" "${TMP_FILE}"' EXIT
curl -fsSL -o "${TMP_ARCHIVE}" "${URL}"
```

The same change is needed in `cmd/geninstaller/install.sh.tmpl`.

files/install.sh.tmpl
```diff
@@ -61,9 +61,16 @@ function user_install() {
...
+  if [ ! -f "${TMP_ARCHIVE}" ] || [ ! -s "${TMP_ARCHIVE}" ]; then
+    echo "Failed to download ${URL}" >&2
+    exit 1
+  fi
+  gzip -dc "${TMP_ARCHIVE}" > "${TMP_FILE}"
+  rm -f "${TMP_ARCHIVE}"
```
alecthomas, line 73: 🤖 Temp files are left behind on both failure paths. The `exit 1` for a missing or empty archive never removes `TMP_ARCHIVE`, and a `gzip` failure aborts via `set -e` before the `rm -f` on line 73.

Observed against a fake `file://` release:

- Zero-byte download: leaves `hermit.download.20463.gz` in `pkg/hermit@canary/`.
- Corrupt (non-gzip) archive: leaves both `hermit.download.27248.gz` and the empty `hermit.download.27248`.

The decompressed leftover pre-dates this PR, but the `.gz` leftover is new, and cleanup is one of the stated goals here. Every retry adds more debris to the package directory. The `trap` suggested above handles all of these cases in one line, and also removes the need for `--remove-on-error`.

Same applies to `cmd/geninstaller/install.sh.tmpl`.

files/install.sh.tmpl
```diff
@@ -61,9 +61,16 @@ function user_install() {
   fi
   URL="${HERMIT_DIST_URL}/hermit-${OS}-${ARCH}.gz"
   TMP_FILE="${HERMIT_EXE}.download.${RANDOM}"
+  TMP_ARCHIVE="${TMP_FILE}.gz"
   echo "Downloading ${URL} to ${HERMIT_EXE}"
   chmod -f u+w "${HERMIT_EXE}" 2> /dev/null || true
-  curl -fsSL "${URL}" | gzip -dc > "${TMP_FILE}"
+  curl --remove-on-error -fsSL -o "${TMP_ARCHIVE}" "${URL}"
+  if [ ! -f "${TMP_ARCHIVE}" ] || [ ! -s "${TMP_ARCHIVE}" ]; then
```
alecthomas, line 68: 🤖 Minor: this existence test is redundant. `-s` is already false when the file does not exist, so `[ ! -s "${TMP_ARCHIVE}" ]` alone covers both conditions, including the case where a `curlrc` `output-dir` setting redirects the file elsewhere.

alecthomas (changes requested):
🤖 The underlying fix is correct, and I confirmed it: with a `curlrc` containing `write-out`, the previous script fails (`gzip: (stdin): trailing garbage ignored`, exit 2) while this version installs cleanly. Both templates stay byte-identical and the new lines are shellcheck-clean.

One blocking problem, though. `--remove-on-error` was added in curl 7.83.0 (May 2022), and curl exits with code 2 on an unknown option before making any request, so under `set -euo pipefail` the entire installation aborts on any older curl. Verified with a stub emulating curl 7.81: the previous script installs successfully, this one fails with `curl: option --remove-on-error: is unknown` and installs nothing. That covers Ubuntu 22.04 LTS (7.81.0), Ubuntu 20.04 (7.68.0), Debian 11 (7.74.0), RHEL/Rocky 8 (7.61.1) and RHEL 9 (7.76.1), and macOS 12 and earlier. It breaks both the documented `curl ... install.sh | bash` bootstrap and the self-healing path in every generated `bin/hermit` stub, and CI cannot catch it because `ubuntu-latest` and `macos-latest` ship curl 8.x. The flag is not load-bearing either, since `-f` plus `set -e` already aborts on a failed download; a `trap 'rm -f "${TMP_ARCHIVE}" "${TMP_FILE}"' EXIT` is portable and also cleans up the temp files that are currently left behind when the archive is empty or not valid gzip.

Details, including the redundant `-f` test, are in the inline comments. Separately, the validation harnesses described in the PR body (the `curlrc` regression case and the zero-byte rejection) are not in the diff, and no existing test exercises either path, so this regression can return silently.

PR #605, opened 2026-08-12
Add GitHub Enterprise Cloud host support

mightyguava:
## Summary

Add host-aware GitHub Enterprise support for private release downloads, HTTPS git source rewrites, and manifest inference.

This primarily supports GitHub Enterprise Cloud with data residency, where the web host is `<subdomain>.ghe.com` and the REST API host is `api.<subdomain>.ghe.com`. Additional GitHub-compatible hosts can be configured explicitly when they expose the same `https://api.<host>` REST layout.

### 1. Repeatable, host-aware authentication config

`github-token-auth` is now repeatable and accepts `host`, `token-env`, and `match`:

```hcl
github-token-auth {
  host = "mycompany.ghe.com"
  token-env = "HERMIT_GITHUB_TOKEN_MYCOMPANY_GHE_COM"
  match = ["mycompany/*"]
}
```

The host defaults to `github.com`. An explicit block makes its `match` globs authoritative, so nonmatching repositories fall through to the normal package source.

For GHEC data residency hosts matching the single-subdomain `<subdomain>.ghe.com` form, the block is optional for release downloads. Hermit recognizes those hosts automatically and lazily resolves credentials with:

```sh
gh auth token -h <host>
```

Other enterprise hosts require an explicit block.

### 2. One explicit multi-host GitHub API

`github.HostConfig` configures each web host and token. A single client is constructed with:

```go
github.New(client, []github.HostConfig{...})
```

The client exposes one host-explicit method set instead of legacy and `ForHost` variants:

- `Repo(host, repo)`
- `Release(host, repo, tag)`
- `LatestRelease(host, repo)`
- `Releases(host, repo, limit)`
- `Download(host, asset)`
- `ETag(host, asset)`
- `ProjectForURL(sourceURL) (host, project)`

The REST API base is derived as `https://api.<host>` and uses the existing REST v3 endpoint paths. There is no configurable API path or base URL.

### 3. Host-scoped token transport and CLI fallback

The transport API is consolidated to:

```go
github.TokenAuthenticatedTransport(transport, []github.HostConfig{...})
```

A token is only sent to its configured web host and corresponding `api.<host>`. It is never sent to GitHub.com or unrelated redirect hosts.

Configured enterprise hosts resolve their token before constructing both the HTTP client and HTTPS git source rewriter, so the documented `gh` fallback works for both paths. Automatically recognized, unconfigured GHEC release hosts resolve the token lazily when requested.

CLI lookups are cached per host with `sync.OnceValues`. The subprocess runs outside the shared mutex, failures are cached, and lookups for different hosts do not block each other.

### 4. Host-aware private release selection

A single aggregate selector recognizes GitHub-style release URLs:

```text
https://<host>/<owner>/<repo>/releases/download/<tag>/<asset>
```

It supports:

- `github.com` when matched by configuration
- single-subdomain `*.ghe.com` hosts automatically
- explicitly configured GitHub-compatible hosts

When matched, Hermit queries the host-specific release API and downloads the selected asset instead of following browser login or SSO redirects.

Configured matchers take precedence over automatic GHEC recognition. Release URLs containing a query or fragment are left to the normal source path, preserving the previous anchored URL behavior.

### 5. Host normalization and redirect safety

Host comparison is normalized across configuration, artifact URLs, API requests, asset validation, and git rewrites. Normalization handles case, schemes, trailing slashes, and an explicit default `:443` port.

Only `github.com` and the single-subdomain `<subdomain>.ghe.com` form are recognized automatically. Object storage hosts such as `objects-origin.<subdomain>.ghe.com` are not treated as GitHub hosts and do not trigger token lookup.

Release asset and ETag requests validate that the API asset URL belongs to the selected web/API host. Enterprise downloads also reject redirects to `github.com` or `api.github.com`. Presigned redirects to object storage remain allowed and receive no GitHub token.

### 6. HTTPS git source rewriting

`AuthenticatedURLRewriter(host, token, matcher)` rewrites matched HTTPS git sources while preserving the enterprise host:

```text
https://mycompany.ghe.com/owner/repo.git
```

becomes:

```text
https://x-access-token:<token>@mycompany.ghe.com/owner/repo.git
```

SSH URLs remain unchanged.

### 7. Credential redaction

The implementation is integrated with the current `redact.Secret` and `redact.URL` APIs. Tokens and credential-bearing rewritten URLs are revealed only at their deliberate use points, preventing credentials from appearing in normal logs and errors.

### 8. Host-aware manifest inference

Manifest inference carries the detected host into `ProjectForURL` and `Repo`, so enterprise artifact URLs query `https://api.<host>` instead of `api.github.com`.

For non-`github.com` hosts, inference does not emit `github-release = "owner/repo"` because enterprise autoversion remains out of scope.

## Notes

- Autoversion through `github-release` remains GitHub.com-only and explicitly passes `github.com` to the host-aware client.
- GHEC data residency release handling is intentionally automatic for single-subdomain `*.ghe.com` hosts.
- Explicit `github-token-auth` blocks scope handling with `match` and can override `token-env`.
- Self-hosted or custom domains must be configured and must expose the supported `https://api.<host>` REST layout.
- `gh auth status --json hosts` is not used for discovery. Host recognition is deterministic: `github.com`, single-subdomain `*.ghe.com`, and explicit configuration.

## Test Plan

- `go test ./...`
- `go test -race ./github`
- `golangci-lint run`
- `just docs-schema`
- `git diff --check`


app/main.go
```diff
@@ -352,3 +347,57 @@ func fatalIfError(logger *ui.UI, ctx *kong.Context, err error) {
...
+			p.Tracef("GitHub token for %s set from %s", host.WebHost, source)
+		}
+		auths = append(auths, gitHubAuth{host: host, matcher: matcher})
+	}
+	return auths, nil
+}
+
+func githubTokenForHost(config hermit.GitHubTokenAuthConfig) (token, source string) {
+	if config.TokenEnv != "" {
+		token := os.Getenv(config.TokenEnv)
```
alecthomas: **P0 — CI failure.** `golangci-lint run` reports `shadow: declaration of "token" shadows declaration at line 386 (govet)` here and again at line 398. The named return value `token` on `githubTokenForHost` is shadowed by both inner `token` declarations.

I ran the linter against `master` to be sure this is new: the baseline only reports two pre-existing findings, both from a stale cache path outside the repository, so these are genuinely introduced by this branch and `.github/workflows/ci.yml` will fail.

Dropping the named returns, or renaming the inner variables, resolves both.

github/api.go
```diff
@@ -193,3 +363,11 @@ func (a *Client) request(method string, url string, headers http.Header) (*http.
 	req.Header = headers
 	return req, nil
 }
+
+var ghAuthToken = func(host string) (string, error) {
+	out, err := exec.Command("gh", "auth", "token", "-h", host).Output() //nolint:noctx
```
alecthomas: **P0 — CI failure.** `golangci-lint run` reports `directive //nolint:noctx is unused for linter "noctx" (nolintlint)`. The `noctx` linter does not flag `exec.Command`, so the suppression has nothing to suppress and `nolintlint` rejects it. Removing the directive is enough.

cache/github.go
```diff
@@ -5,28 +5,52 @@ import (
...
-		if !ok || match == nil || !match(info.owner, info.repo) {
+		if !ok || !ghclient.SupportsHost(info.host) || !matchesGitHubHost(info, normalizedMatchers) {
 			return getSource(client, uri)
 		}
 		return &githubReleaseSource{url: uri, info: info, ghclient: ghclient}, nil
 	}
 }
 
+func matchesGitHubHost(info *githubReleaseInfo, matchers []GitHubHostMatcher) bool {
+	if info.host != "github.com" && strings.HasSuffix(info.host, ".ghe.com") {
```
alecthomas: **P1 — the `match` globs are silently ignored for every `.ghe.com` host.**

This early return fires before the matcher loop below is ever reached, so for a host such as `mycompany.ghe.com` the configured `match = ["mycompany/*"]` has no effect at all. A URL like `https://mycompany.ghe.com/some-other-org/tool/releases/download/v1/tool.tgz` is routed through the authenticated GitHub client and the enterprise token is used for it, even though the user explicitly scoped authentication to `mycompany/*`.

That contradicts what `docs/docs/usage/config.md` documents `match` to mean, and it is inconsistent with `github/url.go`, where the git URL rewriter does honour the matcher for the same host.

I would remove this branch entirely and let the loop below decide, which also makes the two code paths agree.

app/main.go
```diff
@@ -227,24 +224,22 @@ func Main(config Config) {
...
+	}
+	ghClient := github.NewWithHosts(defaultHTTPClient, githubHosts)
 
-			getSource = cache.GitHubSourceSelector(getSource, ghClient, matcher)
-		}
+	var githubHostMatchers []cache.GitHubHostMatcher
+	for _, auth := range githubAuths {
+		githubHostMatchers = append(githubHostMatchers, cache.GitHubHostMatcher{Host: auth.host.WebHost, Match: auth.matcher})
 	}
+	getSource = cache.GitHubSourceSelectorForHosts(getSource, ghClient, githubHostMatchers)
```
alecthomas, line 246: **P1 — enterprise handling is now always on, and any `.ghe.com` URL can trigger a `gh` subprocess.**

The selector is installed unconditionally here, whereas the previous code installed it only when `github-token-auth.match` was non-empty. Combined with `IsGitHubHost` treating every three-label `*.ghe.com` name as GitHub-compatible, and with the unconditional `.ghe.com` branch in `matchesGitHubHost`, this means any package manifest whose artefact URL looks like `https://<anything>.ghe.com/o/r/releases/download/t/a` is routed into the authenticated client. That client then shells out to `gh auth token -h <host>`, and `github/http.go` turns a `gh` failure into a `RoundTrip` error, so the download fails outright.

There are two consequences. First, it is a regression: a `.ghe.com` artefact that previously downloaded over plain HTTPS now hard-fails on any machine without an authenticated `gh`. Second, it means a remote, manifest-supplied URL causes Hermit to execute a `gh` binary resolved from `PATH`. The token itself only ever travels to the host it belongs to, so this is not a credential-exfiltration bug, but running a subprocess in response to untrusted input deserves an explicit opt-in.

The behaviour is asserted as intentional by `TestGitHubSourceSelectorRecognizesGHEByDefault`, but it is not mentioned anywhere in the documentation this PR updates, which tells users to configure `host` and `token-env` explicitly. I would drop the `.ghe.com` special case and require the host to be configured.

app/main.go
```diff
@@ -281,8 +276,8 @@ func Main(config Config) {
 
 	var sourceRewriters []sources.URLRewriter
 	if isActivated {
-		if matcher != nil {
-			sourceRewriters = append(sourceRewriters, github.AuthenticatedURLRewriter(githubToken, matcher))
+		for _, auth := range githubAuths {
+			sourceRewriters = append(sourceRewriters, github.AuthenticatedURLRewriterForHost(auth.host.WebHost, auth.host.Token, auth.matcher))
```
alecthomas: **P1 — the documented `gh` fallback does not apply to git source URLs.**

`docs/docs/packaging/private.md` tells users that for non-GitHub.com hosts they may "set `token-env` or authenticate the GitHub CLI for the host". That holds for the HTTP transport, but not for source rewriting.

`auth.host.Token` is empty whenever `githubTokenForHost` sees a non-GitHub.com host with no `token-env` set, and `github/url.go` returns the URL unchanged when the token is empty. So a user who relies on `gh` for their enterprise host will find that artefact downloads authenticate correctly but `sources` clones silently do not, and they will get an opaque git failure rather than a clear error.

Either resolve the `gh` token eagerly here so the rewriter receives it, or fail loudly when an enterprise host has neither `token-env` nor a resolvable `gh` token.

cache/github_test.go
```diff
@@ -0,0 +1,65 @@
...
+type testPackageSource struct{}
+
+func (testPackageSource) OpenLocal(*Cache, string) (*os.File, error) { return nil, nil }
+func (testPackageSource) Download(*ui.Task, *Cache, string) (string, string, string, error) {
+	return "", "", "", nil
+}
+func (testPackageSource) ETag(*ui.Task) (string, error) { return "", nil }
+func (testPackageSource) Validate() error               { return nil }
+
+func TestGitHubSourceSelector(t *testing.T) {
```
alecthomas, line 23: **P2 — this test passes for the wrong reason.** The matcher is `mycompany/*` and the URL under test is `mycompany/tool`, so the assertion holds whether or not the matcher is consulted. Because of the unconditional `.ghe.com` branch in `matchesGitHubHost`, the test would still pass if the glob were `nomatch/*`.

A case asserting that a *non-matching* repo on a configured `.ghe.com` host falls through to the plain package source would fail today, and is the case worth having.

docs/docs/usage/user-config-schema.hcl
```diff
@@ -29,7 +29,12 @@ defaults {
 
   # When to use GitHub token authentication.
   github-token-auth {
-    # One or more glob patterns. If any of these match the 'owner/repo' pair of a GitHub repository, the GitHub token from the current environment will be used to fetch their artifacts.
+    # GitHub web host for matched repositories. Defaults to github.com. For GitHub Enterprise Cloud with data residency this is typically <subdomain>.ghe.com.
+    # default: github.com
+    host = string # (optional)
```
alecthomas, line 36: **P2 — this generated file is out of sync.** It looks hand-edited rather than regenerated. Running `just docs-schema` emits `github-token-auth { # (repeated)` on line 31, reflecting the block becoming a slice in `env.go`.

(There is also a missing `install-on-activate` entry, but I confirmed that drift pre-dates this PR.)

cache/github.go
```diff
@@ -78,42 +102,45 @@ func (g *githubReleaseSource) getAsset() (github.Asset, error) {
...
 type githubReleaseInfo struct {
-	owner, repo, tag, asset string
+	host, owner, repo, tag, asset string
 }
 
 func getGitHubReleaseInfo(uri string) (*githubReleaseInfo, bool) {
-	g := &githubReleaseInfo{}
-	m := githubRe.FindStringSubmatch(uri)
-	if len(m) != 5 {
+	u, err := url.Parse(uri)
```
alecthomas, line 127: **P2 — release URLs with a query or fragment are now accepted.** The previous anchored regex required the URL to end at the asset name. This version only inspects `u.Path`, so `.../releases/download/v1/tool.tgz?sig=...` is now treated as a release asset, re-fetched through the API, and the query string is silently discarded.

If any manifest carries a signed or parameterised release URL, this changes what actually gets downloaded. Rejecting URLs with a non-empty `RawQuery` or `Fragment` would preserve the old behaviour.

cache/github.go
```diff
@@ -5,28 +5,52 @@ import (
...
 		return &githubReleaseSource{url: uri, info: info, ghclient: ghclient}, nil
 	}
 }
 
+func matchesGitHubHost(info *githubReleaseInfo, matchers []GitHubHostMatcher) bool {
+	if info.host != "github.com" && strings.HasSuffix(info.host, ".ghe.com") {
+		return true
+	}
+	for _, matcher := range matchers {
+		if matcher.Host == info.host && matcher.Match != nil && matcher.Match(info.owner, info.repo) {
```
alecthomas: **P2 — host comparison is not normalised.** `info.host` is the raw `u.Host` taken from the URL, while `matcher.Host` has already been through `NormalizeHost`. An uppercase host, or one carrying an explicit `:443`, will therefore fail to match.

There is also a predicate mismatch worth tidying: `IsGitHubHost` requires exactly three labels, whereas the branch above uses a bare `.ghe.com` suffix check. The two should share one implementation.

github/http.go
```diff
@@ -2,29 +2,106 @@ package github
...
+	if token := g.tokensByWebHost[webHost]; token != "" {
+		return token, nil
+	}
+	if webHost == gitHubHost {
+		return "", nil
+	}
+	if token := g.ghTokens[webHost]; token != "" {
+		return token, nil
+	}
+	token, err := ghAuthToken(webHost)
```
alecthomas: **P2 — `ghAuthToken` runs while holding the mutex.** `g.mu` is taken above and then a subprocess is forked here, which serialises every concurrent GitHub request behind the `gh` invocation. Failures are also not recorded, so each subsequent request re-runs the subprocess after a failure.

Caching the outcome (including the error) and releasing the lock around the exec would avoid both.

github/api.go
```diff
@@ -44,67 +81,132 @@ type Asset struct {
...
-		client = &http.Client{Transport: TokenAuthenticatedTransport(client.Transport, token)}
+	if len(hosts) == 0 {
+		hosts = []HostConfig{{WebHost: gitHubHost}}
+	}
+
+	normalized := make(map[string]HostConfig, len(hosts))
+	for i, host := range hosts {
+		host.WebHost = NormalizeHost(host.WebHost)
+		normalized[host.WebHost] = host
+		hosts[i] = host
```
alecthomas: **P2 — this mutates the caller's slice.** `NewWithHosts` writes normalised entries back into the `hosts` argument. Nothing depends on it today, but it is a surprising side effect on an exported constructor, and the normalised values are already being collected into `normalized` on the line above.

app/main.go
```diff
@@ -352,3 +347,57 @@ func fatalIfError(logger *ui.UI, ctx *kong.Context, err error) {
...
+		host.WebHost = github.NormalizeHost(host.WebHost)
+		if source != "" {
+			p.Tracef("GitHub token for %s set from %s", host.WebHost, source)
+		}
+		auths = append(auths, gitHubAuth{host: host, matcher: matcher})
+	}
+	return auths, nil
+}
+
+func githubTokenForHost(config hermit.GitHubTokenAuthConfig) (token, source string) {
```
alecthomas: **P2 — no tests.** `githubTokenForHost` and `configuredGitHubAuths` are new logic covering environment-variable precedence, per-host fallback, and host normalisation, and neither has a test, even though the `app` package has a test suite.

`github.NormalizeHost` is also untested despite every host comparison in this PR depending on it. Its edge cases (ports, uppercase, trailing slash, an explicit scheme prefix) are exactly where the mismatches flagged elsewhere in this review come from.

sources/sources_test.go
```diff
@@ -60,6 +68,9 @@ func TestGitHubTokenRewriter(t *testing.T) {
 			assert.NoError(t, err)
 
 			rewriter := github.AuthenticatedURLRewriter(tt.token, matcher)
+			if tt.name == "matching enterprise github repo" {
```
alecthomas: **P2 — branching on the test name makes the table no longer self-describing.** Adding a `host` field to the table struct, defaulted to `github.com`, would let every row go through the same rewriter construction and would document the enterprise case in the row itself.

github/api.go
```diff
@@ -44,67 +81,132 @@ type Asset struct {
...
 }
 
-// New creates a new GitHub API client.
+// New creates a new GitHub.com API client.
 func New(client *http.Client, token string) *Client {
+	return NewWithHosts(client, []HostConfig{{WebHost: gitHubHost, Token: token}})
+}
+
+// NewWithHosts creates a new GitHub API client for one or more GitHub hosts.
+func NewWithHosts(client *http.Client, hosts []HostConfig) *Client {
```
alecthomas: **P2 — API surface duplication.** Every `Client` method now has an `XForHost` twin, and the host-less variants are unused inside the repository apart from `New`, `GitHubSourceSelector`, and `AuthenticatedURLRewriter`. Threading the host through the existing methods, or having `NewWithHosts` return per-host sub-clients, would avoid doubling the exported surface.

This is a judgement call rather than a defect, but it is where most of the diff's size comes from.

alecthomas (changes requested):

github/api.go
alecthomas: I think it's okay to break API compatibility if it keeps the API cleaner.

app/main.go
mightyguava: Fixed in `d6124f8`: `githubTokenForHost` now uses unnamed returns, eliminating both shadowing findings. `golangci-lint run` passes.

cache/github.go
mightyguava: Fixed in `d6124f8`: an explicit matcher for a host now takes precedence, so nonmatching repositories fall through. Unconfigured recognized GHEC hosts retain the intended automatic handling.

app/main.go
mightyguava: Keeping this behavior as discussed: automatic recognition of single-subdomain `*.ghe.com` hosts is intentional for GHEC data residency. Explicitly configured hosts still honor their `match` scope, and the docs now call out the automatic `gh auth token -h <host>` fallback. No opt-in change planned.

github/api.go
mightyguava: Fixed in `d6124f8`: removed the unused `//nolint:noctx` directive. `golangci-lint run` passes.

app/main.go
mightyguava: Fixed in `d6124f8`: configured enterprise hosts resolve the `gh` token before constructing both the HTTP host config and HTTPS git URL rewriter, so the documented fallback applies to both paths and fails clearly when unavailable.

cache/github_test.go
mightyguava: Fixed in `d6124f8`: added a nonmatching repository case for an explicitly configured `.ghe.com` host and asserted that it returns the fallback package source.

docs/docs/usage/user-config-schema.hcl
mightyguava: Fixed in `d6124f8`: regenerated the schema with `just docs-schema`, including the repeated-block marker and the pre-existing generated schema drift.

cache/github.go
mightyguava: Fixed in `d6124f8`: `getGitHubReleaseInfo` now rejects URLs with a non-empty query or fragment, with regression coverage for both cases.

app/main.go
mightyguava: Addressed in `d6124f8`: added coverage for token environment precedence, enterprise CLI fallback and errors, configured auth construction, and `NormalizeHost` edge cases.

github/api.go
mightyguava: Addressed across `d6124f8` and `e82ffbb`: the client now has one host-explicit method set, `New` accepts `[]HostConfig`, and the transport API is consolidated to `TokenAuthenticatedTransport(transport, []HostConfig)`. The legacy wrappers and `ForHost` variants are removed.

cache/github.go
mightyguava: Fixed in `d6124f8`: release URL hosts and configured hosts now use the same normalization, including lowercase and default `:443` handling, and matching uses `IsGitHubHost` consistently.

github/api.go
mightyguava: Fixed in `d6124f8`: host normalization is collected without writing normalized entries back into the input slice. Added an immutability test for `New`.

sources/sources_test.go
mightyguava: Fixed in `d6124f8`: the table now carries `host` explicitly and every case uses the same rewriter construction path.

github/http.go
mightyguava: Fixed in `d6124f8`: CLI lookups use a per-host `sync.OnceValues` resolver created under the mutex and invoked after unlocking. Both tokens and errors are cached, and concurrent different-host lookup is tested.

PR #608, opened 2026-08-13
fix: Keep GitHub token out of git source URLs

alecthomas:
The token was embedded in the source URL, so it reached every consumer
that treats that URL as an identifier: command lines in logs and error
messages, the UI task label, hermit status output, the source directory
hash, and remote.origin.url in the cloned repository's .git/config,
where it persisted indefinitely and was reused on every pull.

Pass it to git as an http.extraheader via GIT_CONFIG_* environment
variables instead. The environment keeps it out of argv as well as out
of .git/config, so all of those sinks close at once. Redacting URL
userinfo where commands and sources are displayed covers the remaining
case of credentials a user hand-writes into sources.

DX-26

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/app/exec_cmd.go
+++ b/app/exec_cmd.go
@@ -32,7 +32,7 @@ func (e *execCmd) Run(
 	globalState GlobalState,
 	config Config,
 	defaultHTTPClient *http.Client,
-	sourceRewriters []sources.URLRewriter,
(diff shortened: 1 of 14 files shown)


alecthomas:
A correct implementation of this is quite tricky, and the risk is relatively low.

PR #616, opened 2026-08-17
fix: redact credentials from logs and errors

alecthomas:
With github-token-auth configured, the GitHub token was embedded in
rewritten source URLs and leaked in cleartext through task labels,
hermit status, command-line error messages and relayed git output.
Credentials embedded in manifest source URLs leaked the same way.

Wrap sensitive values in redacting types (redact.Secret, redact.URL)
at application boundaries - environment variables, HCL fields and CLI
flags - and carry them through typed APIs, revealing the raw value
only at exec argv, HTTP requests and cache hashing. Subprocess output
is scrubbed by exact match since git echoes tokenised URLs, and
stdlib *url.Error is stripped to its cause so its message cannot
reintroduce the raw URL. Serialisation stays raw so config files and
sha256sums round-trip verbatim, and cache paths still hash the raw
URL so existing caches remain valid.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/app/init_cmd.go
+++ b/app/init_cmd.go
@@ -3,14 +3,15 @@ package app
 import (
 	"github.com/cashapp/hermit"
 	"github.com/cashapp/hermit/errors"
+	"github.com/cashapp/hermit/redact"
 	"github.com/cashapp/hermit/ui"
 )
 
 type initCmd struct {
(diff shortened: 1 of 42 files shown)


shellz-n-stuff (approved):
🔥 /10 nice one

